PR #2, opened 2020-06-17
rewrite build-toolchain.sh

cometodukey:
This patch allows `toolchain/build-toolchain.sh` to build on more systems. It also slightly improves speed, reduces external dependancies and improves readability.

### DIFF
--- a/toolchain/build-toolchain.sh
+++ b/toolchain/build-toolchain.sh
@@ -1,12 +1,17 @@
 #!/usr/bin/env sh
 
+log() {
+	printf "\033[33m->\033[m %s\n" "$*"
+}
+
 download() {
-	wget "$1" -O "$2"
-	tar xvf "$2"
+	log "Downloading $2"
+	wget -c "$1"
+	log "Inflating $2"
+	tar xf "$2"
 }
 
 download_binutils() {
-	printf " - binutils\n"
 	version="2.34"
 	binutils="binutils-$version.tar.xz"
 	download "https://ftp.gnu.org/gnu/binutils/$binutils" "$binutils"
@@ -14,15 +19,14 @@ download_binutils() {
 }
 
 download_gcc() {
-	printf " - gcc\n"
-	version="9.3.0"
+	version="10.1.0"
 	gcc="gcc-$version.tar.xz"
 	download "ftp://ftp.gnu.org/gnu/gcc/gcc-$version/$gcc" "$gcc"
 	export gcc_src="$PWD/gcc-$version"
 }
 
 build_binutils() {
-	printf "Building binutils\n"
+	log "Building binutils"
 	cd binutils
 	"$binutils_src/configure" --target="$TARGET" 	\
 					    	  --prefix="$PREFIX" 	\
@@ -35,7 +39,7 @@ build_binutils() {
 }
 
 build_gcc() {
-	printf "Building gcc\n"
+	log "Building gcc"
 	cd gcc
 	"$gcc_src/configure" --target="$TARGET"			\
 					 	 --prefix="$PREFIX" 		\
@@ -66,7 +70,6 @@ main() {
 
 	# download sources
 	cd src
-	printf "Downloading dependancies\n"
 	download_binutils
 	download_gcc
 	cd .. # toolchain/
--- a/toolchain/build-toolchain.sh
+++ b/toolchain/build-toolchain.sh
@@ -1,69 +1,80 @@
-#!/bin/bash
+#!/usr/bin/env sh
 
-MAKE_OPTIONS="-j1"
-
-function download() { # $1=url, $2=file
-	if [ ! -f $2 ]; then
-		curl "$1$2" > "$2"
-	fi
-
-	if [ ! -d "${2%.tar.xz}" ]; then
-		tar xf "$2"
-	fi
+download() {
+	wget "$1" -O "$2"
+	tar xvf "$2"
 }
 
-function download-binutils() {
-	url="ftp://ftp.gnu.org/gnu/binutils/"
-	file=$(curl -sl $url | grep -E 'binutils-[0-9.]+.tar.xz$' | sort -V | tail -n1)
-	download "$url" "$file"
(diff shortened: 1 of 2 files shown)


toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
-#!/bin/bash
+#!/usr/bin/env sh
```
29jm, line 1: I'm only really familiar with bash (`sh` defaults to dash on Debian) so I'm not sure about this. By sticking to the usual features I'm sure it'll work the same, but would I even know what feature is bash-specific or not?

toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
...
-function download() { # $1=url, $2=file
-	if [ ! -f $2 ]; then
-		curl "$1$2" > "$2"
-	fi
-
-	if [ ! -d "${2%.tar.xz}" ]; then
-		tar xf "$2"
-	fi
+download() {
+	wget "$1" -O "$2"
```
29jm: the `-O` option overwrites existing files which may be unnecessary. I think `wget --continue "$1"` would be cool though!

toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
...
-	url="ftp://ftp.gnu.org/gnu/binutils/"
-	file=$(curl -sl $url | grep -E 'binutils-[0-9.]+.tar.xz$' | sort -V | tail -n1)
-	download "$url" "$file"
-	echo "$PWD/${file%.tar.xz}"
+download_binutils() {
+	printf " - binutils\n"
+	version="2.34"
+	binutils="binutils-$version.tar.xz"
+	download "https://ftp.gnu.org/gnu/binutils/$binutils" "$binutils"
+	export binutils_src="$PWD/binutils-$version"
```
29jm, line 18: `export` is a much better solution than what I'd come up with...

toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
...
-	if [ ! -f $2 ]; then
-		curl "$1$2" > "$2"
-	fi
-
-	if [ ! -d "${2%.tar.xz}" ]; then
-		tar xf "$2"
-	fi
+download() {
+	wget "$1" -O "$2"
+	tar xvf "$2"
```
29jm: That little `v` is pretty spammy, maybe a single print like "Extracting $2" ?

toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
...
-make install
-cd ..
+	export PREFIX="$PWD/compiler"
+	export TARGET="i686-elf"
+	export PATH="$PREFIX/bin:$PATH"
 
-echo "Building GCC"
+	# download sources
+	cd src
+	printf "Downloading dependancies\n"
```
29jm: minor typo

toolchain/build-toolchain.sh
```diff
@@ -1,69 +1,80 @@
...
-function download-gcc() {
-	url="ftp://ftp.gnu.org/gnu/gcc/"
-	folder=$(curl -sl $url | grep -E 'gcc-[0-9.]+$' | sort -V | tail -n1)
-	url="$url$folder/"
-	file="$folder.tar.xz"
-	download "$url" "$file"
-	echo "$PWD/$folder"
+download_gcc() {
+	printf " - gcc\n"
+	version="9.3.0"
```
29jm: why not 10.1.0 ?

29jm (commented):
Good rewrite! Some little things here and there. It's cool if you can fix those here, but it's also cool with me to merge as-is and I make the changes afterwards, whatever works with you.  
In any case thank you for the PR!

toolchain/build-toolchain.sh
cometodukey: The entire script is POSIX compliant, and `sh` is symlinked to a POSIX compliant shell on every UNIX-like system that I've ever used. I tend to avoid using `bash` specific code when writing shell scripts to ensure portability. However, I think that `bash` emulates POSIX shell behaviour when invoked as `sh` or with the `--posix` option.

toolchain/build-toolchain.sh
cometodukey: I completely forgot that 10.1.0 existed. I'll change it in the next commit.

toolchain/build-toolchain.sh
cometodukey: Thanks, good catch.

toolchain/build-toolchain.sh
cometodukey: That's true. My fix is this.
```sh
log() {
	printf "\033[33m->\033[m %s\n" "$*"
}

download() {
	log "Downloading $2"
	wget "$1" -O "$2"
	log "Inflating $2"
	tar xf "$2"
}
```
`log` will write a yellow arrow and a message, followed by a new line.
Any thoughts?

toolchain/build-toolchain.sh
29jm: I get that, the thing is I'm bound to modify this script at some later point, and I personally won't know what's POSIX compliant or not, so that's a potential source of problem, as opposed to sticking to bash. But that's not really an immediate concern, yeah.

toolchain/build-toolchain.sh
29jm: Looks great!

toolchain/build-toolchain.sh
cometodukey: Ahh, I see. To that end, this seems to be a good resource. https://mywiki.wooledge.org/Bashism. The pure-sh-bible is also very handy. https://github.com/dylanaraps/pure-sh-bible

toolchain/build-toolchain.sh
cometodukey: That looks good to me! I've tested it and it all seems to be working correctly

29jm:
Perfection, thanks again!

PR #5, opened 2020-11-07
mlfq scheduler

nisarg14:


kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
+#include <kernel/sched_robin.h>
+#include <kernel/sys.h>
+#include <kernel/timer.h>
+#include <stdlib.h>
+
+uint32_t flag;
```
29jm, line 6: Put this in `sched_mlfq_t` too

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+
+        kfree(to_remove);
+    }
+}
+
+/* Allocates a mlfq with round robin scheduler.
+ */
+sched_t* sched_mlfq() {
+    sched_mlfq_t* sched = kmalloc(sizeof(sched_mlfq_t));
+    if (timer_get_tick() % 5 != 0)
```
29jm, line 277: Don't do that here, this function will only be called once, in `init_proc`.

You want to have only one `sched_next` function, and have that function check `timer_get_tick()`.

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+}
+
+void sched_mlfq_exit(sched_t* sched, process_t* process) {
+    sched_mlfq_t* sc = (sched_mlfq_t*) sched;
+    proc_node_t* p = sc->processes1;
+    proc_node_t* p2 = sc->processes2;
+    proc_node_t* p3 = sc->processes3;
+
+    
+    if(flag=1){  //exit for processes in circular list 2
```
29jm: You want `==` here and in subsequent ifs

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+		p = p->next;
+	} while (p != sc->processes1);
+
+    sc->processes1 = sc->processes1->next;
+
+    return sc->processes1->process;
+}
+
+//next process for circular list 2
+process_t* sched_mlfq_next2(sched_t* sched) {
```
29jm, line 145: See the comment below on `sched_mlfq`, you want to have only one of those

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+        {        
+            sched->sched = (sched_t) 
+            {
+                .sched_get_current = sched_mlfq_get_current,
+                .sched_add = sched_mlfq_add,
+                .sched_next = sched_mlfq_next,
+                .sched_exit = sched_mlfq_exit
+            };
+            sched->processes1 = NULL;
+            if(sched_mlfq_get_current){
```
29jm, line 289: Function calls need parenthesis; this condition will always be true.

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+        sc->processes1->count = 1; //count 1 for first process
+    } else {
+        proc_node_t* p = sc->processes1->next;
+        sc->processes1->next = new;
+        new->next = p;
+        sc->processes1->count++; //increase count value as process add
+    }
+    
+}
+void sched_mlfq_add_priority(sched_t* sched,process_t* prio_process){
```
29jm, line 65: I'm guessing this is for later, to be used as your `sched_mlfq_t::sched_add` ?

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,328 @@
...
+
+/* The round robin scheduler is simple and requires only a single circular list
+ * containing candidate processes. By having a `sched_t` as the first member of
+ * the struct, we allow casting `sched_robin_t*`s to `sched_t*`.
+ */
+typedef struct {
+    sched_t sched;
+    proc_node_t* processes1;
+    proc_node_t* processes2;
+    proc_node_t* processes3;
```
29jm, line 23: Do circular lists really make sense for this type of scheduler? Could regular lists be used?

29jm (commented):
Alright that's a good start! The most pressing issue is unifying your `sched_add`s.

There are style issues and a bunch of noisy changes (the permission changes) that'll have to be addressed but for now it doesn't matter.

kernel/src/sys/sched_mlfq.c
nisarg14: ` sched_mlfq_add` function is used for adding for first priority queue and this `sched_mlfq_add_priority` is used for addin in queue 2 and 3. Here when process enters it is in first queue , for that we use `sched_mlfq_add` and when it is allowed to execute and after given time quantum if the process is not completed then we send this process to queue 2 and for that we use `sched_mlfq_add_priority`

kernel/src/sys/sched_mlfq.c
nisarg14: Here I have used `timer_get_tick()` as a timer for round robin

kernel/src/sys/sched_mlfq.c
```diff
@@ -231,7 +232,7 @@ void sched_mlfq_exit(sched_t* sched, process_t* process) {
     proc_node_t* p3 = sc->processes3;
 
     
-    if(flag=1){  //exit for processes in circular list 2
+    if(flag==1){  //exit for processes in circular list 2
```
29jm, line 235: Use `sc->flag` now, and remove the global `flag`.

kernel/src/sys/sched_mlfq.c
29jm: What I meant was that `sched_mlfq_add_priority` isn't called from anywhere right now. I was suggesting that maybe you intended to use it later when things shape up.

kernel/src/sys/sched_mlfq.c
29jm: You misunderstood, `sched_mlfq` is literally called only once in the whole kernel code, by the line you've added in "proc.c". Its purpose is to build the scheduler, its initialization. Adding timer code here doesn't make sense, because that's something you want to do periodically, i.e. in the `sched_next` you've defined. Having changed that, you'll probably find other things to reorganise in your code.

kernel/src/sys/sched_mlfq.c
29jm: Still relevant, there is a problem here.

kernel/src/sys/sched_mlfq.c
```diff
@@ -0,0 +1,329 @@
...
+    {   
+        for (size_t i = 0; i < sched->processes2->count; i++)
+        {        
+            sched->sched = (sched_t) 
+            {
+                .sched_get_current = sched_mlfq_get_current,
+                .sched_add = sched_mlfq_add,
+                .sched_next = sched_mlfq_next2,
+                .sched_exit = sched_mlfq_exit
+            };
```
29jm, line 305: In any case, reassigning scheduler callbacks this way (or at all after initialization) can and must be avoided.

29jm (changes requested):

29jm:
Actually closed this by mistake by deleting a branch. Feel free to open a new PR if you make progress on this front!

Issue #6, opened 2020-11-10
&> in Makefile creates race condition on build


the-grue:
Commit 1b3b3b7 introduces a build race condition.  I believe you meant to use 2> instead of &>.  The & puts the process into the background allowing the next steps to complete before dd/mkfs.ext2/grub-mkrescue completes resulting in all kinds of errors during build and these errors in the log file and a black screen in the emulator if it gets that far:

[ext2.c] invalid signature: 0
[ext2.c] bad superblock, aborting
[paging.c] page fault caused by instruction at 0xC010B2D5 from process 0:
[paging.c] the page at 0xC3FC040C wasn't present 
[paging.c] when a process tried to read from it
[paging.c] this process was in kernel mode
[stacktrace.c] stacktrace:
[stacktrace.c]  0xC010041C: paging_fault_handler
[stacktrace.c]  0xC010A32F: isr_handler
[stacktrace.c]  0xC010A6AF: isr48
[stacktrace.c]  0xC01024C5: ext2_read_block
[stacktrace.c]  0xC010275D: ext2_get_inode
[stacktrace.c]  0xC0103BE4: ext2_read
[stacktrace.c]  0xC0103DE4: ext2_readdir
[stacktrace.c]  0xC0102881: ext2_open
[stacktrace.c]  0xC0101E55: fs_open
[stacktrace.c]  0xC0108DA8: proc_exec
[stacktrace.c]  0xC0101019: kernel_main
[stacktrace.c]  0xC010A4C8: _start
Kernel Panic: abort()

Jim

29jm:
Fixed in #7, thanks again!

PR #7, opened 2020-11-10
Fix race condition in Makefile

the-grue:
Correct race condition where &> is replaced 2>.

See  #6


29jm:
Wow good catch, thanks! I did run into this too, strangely just once or twice in at least a hundred builds. That would have started haunting me ^^

PR #13, opened 2020-11-21
Better mouse cursor

the-grue:
Created arrow shaped mouse cursor and made some changes to resolve streaking in the mouse rectangle along with a disappearing cursor issue I discovered.

Some minor optimizations to replace divisions with shifts were also added to the code.

### DIFF
--- a/kernel/src/misc/wm/wm.c
+++ b/kernel/src/misc/wm/wm.c
@@ -8,7 +8,7 @@
 #include <list.h>
 #include <stdlib.h>
 
-#define MOUSE_SIZE 3
+#define MOUSE_SIZE 16
 
 void wm_draw_window(wm_window_t* win, rect_t rect);
 void wm_partial_draw_window(wm_window_t* win, rect_t rect);
@@ -297,8 +297,10 @@ void wm_draw_window(wm_window_t* win, rect_t rect) {
     kfree(clip_windows);
 
     // Clip the mouse cursor too
-    rect_t mouse_rect = wm_mouse_to_rect(mouse);
-    rect_subtract_clip_rect(&clip_rects, mouse_rect);
+    // Somehow the clipping part causes the cursor to streak
+    // but without it, the cursor disappears every second if no movement...
+//    rect_t mouse_rect = wm_mouse_to_rect(mouse);
+//    rect_subtract_clip_rect(&clip_rects, mouse_rect);
 
     // Draw what's left
     rect_t* clip;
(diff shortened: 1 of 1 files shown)


kernel/src/misc/wm/wm.c
```diff
@@ -435,14 +435,25 @@ wm_window_t* wm_window_at(int32_t x, int32_t y) {
 }
 
 void wm_draw_mouse(rect_t old, rect_t new) {
-    wm_refresh_partial(old);
+
+
+    if((old.top != new.top) || (old.left != new.left))
+    	wm_refresh_partial(old);
```
29jm: Remove the whole `if` and refresh call (leave no blank lines!), and add a call to `wm_refresh_partial(prev)` above [this line](https://github.com/the-grue/SnowflakeOS/blob/master/kernel/src/misc/wm/wm.c#L551). This avoids the extra `if`, which is already done there in that `wm_mouse_callback` function. Also could you remove the `old` argument to `wm_draw_mouse` in the process?

kernel/src/misc/wm/wm.c
```diff
@@ -435,14 +435,25 @@ wm_window_t* wm_window_at(int32_t x, int32_t y) {
...
+        memset((void*) addr, 127, (y+1)*fb.bpp>>3);
         addr += fb.pitch;
     }
+
+    addr += 4*fb.bpp>>3;
+
+    for (int32_t y = 0; y < 6; y++) {
+        memset((void*) addr, 127, 3*fb.bpp>>3);
+	addr += fb.bpp>>3;
+	addr += fb.pitch;
```
29jm: About using a shift, I checked and on -O3 optimisation level, gcc produces the same `shr`-based code as with a division by 8. So all things being equal, I'd rather keep using `fb.bpp/8` for clarity.

Also there's a space/tabs thing going on here, could you make sure indentation is four spaces?

29jm (commented):
It works well on my end, that's a neat looking cursor! A couple minor bureaucratic comments and it's good to go :+1:

kernel/src/misc/wm/wm.c
the-grue: Done and done.

kernel/src/misc/wm/wm.c
the-grue: You know, I thought about that.  I reverted the shifts as requested and adjusted tabs back to spaces.

the-grue:
Thanks!  I actually got out my graph paper and mapped it out.

One recommendation, in reference to your "bureaucratic" comment.  This is your repo/project.  If you have specific coding preferences or style that you would like us contrubitors to adhere to, that might make a good wiki page.  Tabs vs. Spaces, where to put curly braces, etc. can vary from project to project.  In fact, I believe wars have been fought over them.

I'll try to be mindful of these things going forward.  Thanks for the tips!

29jm:
Thank you for all the changes, that's perfect! I took the liberty of squashing the commits while merging, didn't want to do it via github as it's pretty unclear what it'll do, so it doesn't show as merged here... I'll try something else next time.

You're totally right about the need for contributing guidelines, this _would_ remove a lot of uncertainty and after-the-fact requests, I really need to do that. The way things are right now I never know how much to ask from contributors, and they can't know what to do, it's a mess ^^

Issue #15, opened 2020-11-22
Keyboard events go missing when issued too fast


29jm:
Keyboard input is dropped when keys are pressed too fast and/or at the same time.

Reproduce steps:
+ start typing fast in the terminal app, something like "abcdefghijkl"
+ some letters will surely be missing in the terminal prompt

29jm:
This happens because the window manager stores only the last keyboard event it receives from the keyboard driver, [here](https://github.com/29jm/SnowflakeOS/blob/39c0d6083e9db9f1e33741f97e2a272f8c2ff13a/kernel/src/misc/wm/wm.c#L556-L572), and clients (process that have opened a window) may not poll for events fast enough. A program whose event loop is slow is sure to drop key presses.

What should happen is that the wm should store the keyboard events in a queue, and when a client asks for its events, give out the oldest key pressed in the event struct, popping it from the queue.

It'd be better not to modify `wm_event_t` and instead add a per-window key press queue (`list_t` is good), and fill out the `wm_event_t` struct as described above in `wm_get_event`. Or something like resembling that in some way ^^

xadaemon:
Maybe something like a ringbuff works better here?

29jm:
You're right, a fixed size structure is even better than a list. A ring buffer implementation already exists in [pipe.c](https://github.com/29jm/SnowflakeOS/blob/master/kernel/src/misc/pipe.c), it'll be the occasion to make it an independent an reusable thing.

xadaemon:
I will do the splitting it into a reusable component first, and then if you are up to giving me a few pointers I can then try to solve this issue what do you think?

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

29jm:
@octetd That'd be awesome!

For the first part, the ringbuffer code should end up in the libc (as ringbuffer.c perhaps?), same as the existing list structure. It'll probably be useful to have a `ringbuffer_t ringbuffer_new(uint32_t size)` function that allocates via `malloc`, and functions to read and write n bytes from/to the buffer, get the content size... In "pipe.c", `pipe_fs_t` should then end up with two members, `fs` plus a ringbuffer.

As for fixing this bug, there are choices to make regarding how to communicate those multiple events to userspace
[...]

PR #23, opened 2020-12-22
Libc ringbuff implementation

xadaemon:
I'm opening this PR with the libc ring-buffer mentioned in #15, I intend to submit a later pr fixing #15 itself but for now to allow usage elsewhere I decided to open this PR now.

I did some heavy commenting on the code in the hopes that it can sort of explain itself.


libc/include/ringbuffer.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+
+#define RINGBUFFER_EOD -1
+typedef unsigned long int size_t;
```
29jm: Use `size_t` from stddef.h.

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * ringbuffer type for SnowflakeOS, this structure deals in bytes
+ * as such it is untyped and special care must be taken to
+ * read/write consistent lengths or do some separate bookkeeping
+ */
+typedef struct _ringbuffer {
+    size_t sz; // total size
```
29jm: Use unabbreviated names, `size` here, `unread_data` & others in both files. They'll make comments here redundant.

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * initialize a ringbuffer for use
+ * takes a pointer to a declared ring buffer and initializes it with size sz
+ * it will be allocated in sz * isz
+ * returns: 0 on success -1 on failure, if this fails consider
+ * the ringbuffer pointed at by ref invalid and unsafe to use
+ * errno is likely to have a cause for the failure
+ */
+int ringbuffer_init(ringbuffer_t* ref, size_t sz);
```
29jm: This may sound weird, as it usually would be, but put the function-level comments above the implementation, not the declaration. The reason is, when I'm debugging code, I'm looking at the source, it's useful to have the doc just besides. Could you also use the following style?
```c
/* Initialize ... split at ~80 chars
 * Further explanation ...
 */
```

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,68 @@
...
+ * how much space is available before the
+ * writing more data would overwrite data
+ * that has not been read yet
+ */
+size_t ringbuffer_free(ringbuffer_t* ref);
+
+/**
+ * disposes of the ringbuff freeing it's allocated memory
+ */
+int ringbuffer_dispose(ringbuffer_t* ref);
```
29jm: This function shouldn't return anything.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+    ref->sz = sz;
+    ref->unr_data = 0;
+
+// alloc the needed size for this ringbuff and use kmalloc
+// if this is compiled with the proper flag
+#ifndef _KERNEL_
+    ref->data = (uint8_t*) malloc(sz);
+#else
+    ref->data = (uint8_t*) kmalloc(sz);
+#endif
```
29jm: Don't make the distinction, use `malloc` in both cases. `kmalloc` is useful for decorating calls, but in the `libc` we can certainly do without, making this distinction everywhere isn't practical. Same thing with `free`/`kfree`.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    return ref;
+}
+
+size_t ringbuffer_used(ringbuffer_t* ref) {
+    return ref->unr_data;
+}
+
+size_t ringbuffer_free(ringbuffer_t* ref) {
```
29jm: The naming is a bit unfortunate here. I'd suggest `ringbuffer_available_space`, and renaming `ringbuffer_dispose` to `ringbuffer_free`.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+    bool wraps = (ref->w_pos + n) > ref->sz;
+    uint8_t* pos = ref->data + ref->w_pos;
+
+    if (ref->w_pos + n == ref->r_pos) {
+        return -1;
+    }
+
+    for (size_t i = 0; i < n; i++) {
+        uint8_t* i_pos = ref->data + i;
+        *i_pos = *(data + i);
```
29jm: Use `ref->data[i] = data[i]`.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+#ifndef __KERNEL__
+    free(ref->data);
+    free(ref);
+#else
+    kfree(ref->data);
+    kfree(ref);
+#endif
+}
+
+int ringbuffer_write(ringbuffer_t* ref, size_t n, uint8_t* data) {
```
29jm: This function causes buffer overflow when `n` is greater than `ref->sz` (which is a valid use case). You can check that it does with
```c
char str[] = "hello I'm about to be overwritten";
uint8_t buf[10];
uint8_t buf2[100];
ringbuffer_t* rb = ringbuffer_new(10);
rb->data = buf; // it'd do the same with the malloc'd buffer, just less visibly so
ringbuffer_write(rb, 100, buf2);
printf("str is now: %s\n", str);
```

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+
+    for (size_t i = 0; i < n; i++) {
+        uint8_t* i_pos = ref->data + i;
+        *i_pos = *(data + i);
+    }
+    ref->w_pos = ((ref->w_pos + n) % ref->sz) == 0 ? n : (ref->w_pos + n);
+    ref->unr_data = ref->unr_data + n;
+    // returns 0 if the data did fit, 1 if old data was overwritten or
+    // -1 if there is insufficient space before the read pointer to place the data
+    return !wraps ? 0 : 1;
```
29jm: Use defines, like `RINGBUFFER_OK`, `RINGBUFFER_OVERFLOW`...

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+    size_t max = ref->unr_data >= n ? n : ref->unr_data;
+
+    if (ref->unr_data <= 0) {
+        return 0;
+    }
+
+    for (size_t i = 0; i < max; i++) {
+        cp_ptr = (uint8_t*) buffer + i;
+        byte = *(ref->data + (r_car + i));
+        *cp_ptr = byte;
```
29jm: I haven't tried to make it happen, but I'm fairly sure there's a read-past-buffer here, same reason as with the `_write` counterpart. If your read pointer is near the end of the buffer, it should read til the end, then wrap around to the beginning for the remainder of the data to be read.

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+
+#define RINGBUFFER_EOD -1
```
29jm: Unused define.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,97 @@
...
+    free(ref);
+#else
+    kfree(ref->data);
+    kfree(ref);
+#endif
+}
+
+int ringbuffer_write(ringbuffer_t* ref, size_t n, uint8_t* data) {
+    bool wraps = (ref->w_pos + n) > ref->sz;
+    uint8_t* pos = ref->data + ref->w_pos;
```
29jm: Unused variable.

29jm (commented):
Thanks for the PR @octetd ! I had assumed you'd reuse the implementation from pipe.c, but I can dig wanting to reinvent that wheel :) Commented with a few issues I noted & other details.

libc/include/ringbuffer.h
xadaemon: oh I forgot to remove this one I used it shut up gcc in some off tree testing I did to iron out the logic

libc/include/ringbuffer.h
xadaemon: oversight, thanks

libc/src/ringbuffer.c
xadaemon: I will add a guard

libc/src/ringbuffer.c
xadaemon: should I use a define here for the -1 error case?

libc/src/ringbuffer.c
xadaemon: a question that arose while looking more carefully at your example you expect ringbuffer_new to not allocate the buffer?

29jm: No, that was for demonstration purposes. It's hard to show the overflow with malloc'd memory, though it would happen too, as the comment mentions.

libc/src/ringbuffer.c
29jm: Yes, please do! On that subject, I don't get what the error case is with the -1. What's the behavior being prevented here?

libc/src/ringbuffer.c
xadaemon: just an api style question I resolved it already commiting soon

libc/src/ringbuffer.c
xadaemon: now it wont do that anymore I tested and confirmed the bug

libc/src/ringbuffer.c
xadaemon: it would be overwriting non read data as to prevent loss of what is in the buffer already but has not been copied somewhere else

libc/include/ringbuffer.h
xadaemon: I did away with some of the comments as those seemed redundant upon second thought

libc/src/ringbuffer.c
29jm: The `==` comparison is very specific for that though, you probably meant to use an inequality.

I've given it a bit more thought, and I don't think a write should ever fail on a ring buffer. It's the user's responsibility to check whether their write would erase data. No one wants to make a write, then be like, "oh no my data's been erased!". I'll make a new review though instead of crowding this one. Thank you for all the changes already made!

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
+#include <stdbool.h>
```
29jm, line 3: Empty line between pragmas and includes.

libc/include/ringbuffer.h
```diff
@@ -0,0 +1,36 @@
...
+#include <stdint.h>
+#include <stddef.h>
+
+#define RINGBUFFER_OK 0
+#define RINGBUFFER_OVERFLOW 1
+
+
```
29jm: One empty line too many there, and below the struct definition.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+    ref->r_pos = 0;
+    ref->w_pos = 0;
+
+    return 0;
+}
+
+/* Allocate and initialize a ringbuffer returns NULL on failure to allocate */
+ringbuffer_t* ringbuffer_new(size_t size) {
+    ringbuffer_t* ref = malloc(sizeof(ringbuffer_t));
+    if (ringbuffer_init(ref, size) == -1) {
```
29jm, line 32: What if malloc failed and `ref` was null? Then `ringbuffer_init` crashes. Add a blank line above the `if`.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    for (size_t i = 0; i < n; i++) {
+        ref->data[i % ref->size] = data[i];
+    }
+    ref->w_pos = ((ref->w_pos + n) % ref->size) == 0 ? n : (ref->w_pos + n);
+    ref->unread_data = ref->unread_data + n;
+    // returns 0 if the data did fit, 1 if old data was overwritten
+    return !((ref->w_pos + n) > ref->size) ? RINGBUFFER_OK : RINGBUFFER_OVERFLOW;
+}
```
29jm: This function should never fail, and should return the number of bytes written, i.e. `max(n, ref->size)`. When `n < ref->size`, all requested bytes are written. If this meant erasing older unread data, the read pointer is advanced (taking into account wrapping around the end of the buffer) to be one byte after the end of the written bytes. When `n >= ref->size`, it's enough to consider the last `ref->size` bytes from `data` and `memcpy` them into `ref->data`, setting pointers accordingly.

Also, add a blank line after the `for`.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+ * data was available to read also note that buffer is not cleared between reads
+ */
+size_t ringbuffer_read(ringbuffer_t* ref, size_t n, uint8_t* buffer) {
+    uint8_t byte;
+    uint8_t* cp_ptr;
+    size_t r_car = ref->r_pos % ref->size;
+    // read n or unr_data if it is less than n
+    size_t max = ref->unread_data >= n ? n : ref->unread_data;
+
+    if (ref->unread_data <= 0) {
```
29jm: No need for that if, if `unread_data == 0`, the for loop condition won't be met, and the computations later on will keep working if they're sound.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+    free(ref->data);
+    free(ref);
+}
+
+/* Write n bytes to the ringbuffer pointed at by ref returns 0 if the data did
+ * fit, 1 if old data was overwritten or -1 if there is insufficient space before
+ * the read pointer to place the data
+ */
+int ringbuffer_write(ringbuffer_t* ref, size_t n, uint8_t* data) {
+    uint8_t* pos = ref->data + ref->w_pos;
```
29jm: Unused variable. The code should compile with no warnings at all.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/* Read n bytes from the ringbuffer pointed at by ref into block returns n if a
+ * full block is read vallues < n indicate that there is less data than n
+ * available and the returned value was read to buffer a return of 0 means no
+ * data was available to read also note that buffer is not cleared between reads
+ */
+size_t ringbuffer_read(ringbuffer_t* ref, size_t n, uint8_t* buffer) {
+    uint8_t byte;
+    uint8_t* cp_ptr;
```
29jm: Both variables are unused.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+    size_t r_car = ref->r_pos % ref->size;
+    // read n or unr_data if it is less than n
+    size_t max = ref->unread_data >= n ? n : ref->unread_data;
```
29jm: Rename `max` to `bytes_to_read` or something, and remove the comment.

libc/src/ringbuffer.c
```diff
@@ -0,0 +1,96 @@
...
+
+    if (ref->unread_data <= 0) {
+        return 0;
+    }
+
+    for (size_t i = 0; i < max; i++) {
+        buffer[i % ref->size] = ref->data[(r_car + i)];
```
29jm: Extra parentheses in `data[...]`

29jm (commented):
A few more comments. Those are the last stylistic changes I'll ask of you, promise ^^

For the `_write` changes, I'm trying to make sure we have a rock solid data structure here. Avoiding new memory corruption issues is a big deal to me, as those are always very time consuming later on when they inevitably cause issues in unrelated parts of the project. As it is, it's hard to reason about, but I'm sure there's an issue with at least the updating of `unread_data`.

libc/src/ringbuffer.c
xadaemon: found the strict target on make I will comb the code with it those ended leftovers after I changed how the arithmetic is done

libc/src/ringbuffer.c
xadaemon: @29jm  I made a new push now write will return the number of bytes written

29jm:
Just one last thing: `unread_data` is unbounded. Writing 100 bytes in a 10 bytes ringbuffer will tell you that there's -90 bytes of data available. Everything else looks great !

xadaemon:
> Just one last thing: `unread_data` is unbounded. Writing 100 bytes in a 10 bytes ringbuffer will tell you that there's -90 bytes of data available. Everything else looks great !

should be bounded now

29jm:
Would you mind cherry-picking the last commit from the `pr-ringbuffer` branch, or making a commit with similar changes? Ring buffers are hard to get right, I tested thoroughly and found a few more issues:
+ `r_pos` wasn't being updated when data was overwritten; it should always point to the oldest data
+ `unread_data` was indeed bounded, but could now become smaller on writes ^^

I did away with `w_pos`, updating it independently from `r_pos` made things harder, since it's in fact entirely dependent on `r_pos` and `unread_data`.

Merging right after that!

xadaemon:
I might take a little longer on my end today but should be done before long, thanks ^^

-------- Original Message --------

xadaemon:
small update, had a few personal issues, will get it cherry picked tomorrow

xadaemon:
@29jm  I did the cherry-pick as requested ^^

29jm (approved):
Perfection :ok_hand:

29jm:
Thanks a lot!

PR #24, opened 2020-12-30
Fix qemu make

xadaemon:
These tweaks made the qemu make target just work™️ for me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ SYSROOT=$(PWD)/$(SYSROOTDIR)
 INCLUDEDIR=$(SYSROOT)/$(PREFIX)/include
 LIBDIR=$(SYSROOT)/$(PREFIX)/lib
 
-PATH:=$(PATH):$(PWD)/toolchain/compiler/bin
+PATH:=$(PATH):/sbin:$(PWD)/toolchain/compiler/bin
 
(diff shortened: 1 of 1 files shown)


29jm:
It's more that ubuntu (or perhaps debian?) dropped support for sdl rather than qemu (it's still in the docs and still works on other distros), but we can't do anything about that so this is a good change either way. Thanks!

the-grue:
I noticed the sdl issue in Ubuntu 20.04.  Seems there are several changes in the newer distros.

29jm:
@the-grue I don't understand this change though. On archlinux, the latest qemu is still built with sdl2, no patches required. In any case, I only cared about it for its absence-of-ugly-menubar-on-launch, no big deal ^^

the-grue:
@29jm looks like Ubuntu/Debian dropped support for SDL in qemu after 18.04.  This is why I hadn't seen the issue prior to upgrading one machine to 20.04.  Here's an old "Ask Ubuntu" posting about it:

https://askubuntu.com/questions/1146531/qemu-system-i386-sdl-sdl-support-is-disabled-ubuntu-19-04

xadaemon:
I'm using open suse tumbleweed here, sdl support was dropped from qemu here as well

PR #25, opened 2020-12-30
partial fix of #15

xadaemon:
@29jm there goes the pr, in my tests chaging the ringbuffer size at wm.c:70 seems to not make much of a difference and the events are happening there is just this odd delay it seems

### DIFF
--- a/kernel/include/kernel/wm.h
+++ b/kernel/include/kernel/wm.h
@@ -10,6 +10,7 @@
 #include <kernel/uapi/uapi_wm.h>
 
 #define WM_NOT_DRAWN  ((uint32_t) 1 << 31) // Window has _never_ been called wm_render_window
+#define WM_WINEVBUFF_SZ 5
 
 /* ufb: the window's buffer in userspace. Used by the client for drawing
  *  operations. We copy this buffer on request to `kfb`.
--- a/kernel/include/kernel/wm.h
+++ b/kernel/include/kernel/wm.h
@@ -5,6 +5,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <list.h>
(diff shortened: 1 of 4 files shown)


xadaemon:
after adding some debug printf's the window seems to be requesting the events all right but not acting upon them (I will continue the tests on my end here)

xadaemon:
it seems that there is something up with the buffer read or the write because upon dumping the bytes read from the ringbuff also the buffer is not being advanced on write, very strange

xadaemon:
it seems I figured out typing events however something is up with clicks they seems to register only when the mouse is moved and register duplicate

29jm:
Great catch on that for loop! I was just looking at this, it was a puzzling bug for sure. `r_pos` will still have to be updated in case of overflow, is that calculation causing issues too (the one commented out) ?

I don't know if that fixes click issues specifically, but you should replace `|=` assignments with plain `=`, otherwise you're `or`ing a flag with some random value from the stack.

Edit: couldn't tell you how, but I managed to type a smiley character by moving the mouse around (`or`-related probably) ^^ Also, the repeat events lead to really cool visuals in paint!

xadaemon:
I removed all instances of `|=`, ok so I figured a few things I will change ringbuffer around a bit and see if that helps

xadaemon:
@29jm I managed to sort out the bug and make the ringbuffer simpler, last problem is that for some reason the moment you focus back on the terminal(say if you were using calc) it prints out a garbage character

xadaemon:
I'm not losing keystrokes with as low as a 5 items ring buffer can you do some tests too?

xadaemon:
@29jm I managed to fix every issue now should be all good for merging

PR #26, opened 2020-12-31
fix: #15

### LINKED ISSUE #15: Keyboard events go missing when issued too fast

xadaemon:


kernel/include/kernel/wm.h
```diff
@@ -5,10 +5,12 @@
 #include <stdint.h>
 #include <stdbool.h>
 #include <list.h>
+#include <ringbuffer.h>
 
 #include <kernel/uapi/uapi_wm.h>
 
 #define WM_NOT_DRAWN  ((uint32_t) 1 << 31) // Window has _never_ been called wm_render_window
+#define WM_WINEVBUFF_SZ 5
```
29jm: Move this define to wm.c, rename it to WM_EVENT_QUEUE_SIZE

kernel/include/kernel/wm.h
```diff
@@ -22,7 +24,8 @@ typedef struct _wm_window_t {
     int32_t y;
     uint32_t id;
     uint32_t flags;
-    wm_event_t event;
+    // wm_event_t event;
```
29jm: Delete the commented out line

kernel/src/misc/wm/wm.c
```diff
@@ -36,7 +36,16 @@ static uint32_t id_count = 0;
...
 static mouse_t mouse;
 
+void buffer_dump(void* buff, size_t len) {
+    for (size_t i = 0; i < len; i++)
+    {
+        printf("%d:%#2X ", i, ((uint8_t*)buff)[i]);
+    }
+    printf("\n");
+}
+
```
29jm: Remove this, or move it to sys.h (in which case, put the braces on the same line for the for, add a blank line after the for)

kernel/src/misc/wm/wm.c
```diff
@@ -57,7 +66,9 @@ uint32_t wm_open_window(fb_t* buff, uint32_t flags) {
         .ufb = *buff,
         .kfb = *buff,
         .id = ++id_count,
-        .flags = flags | WM_NOT_DRAWN
+        .flags = flags | WM_NOT_DRAWN,
+        /*TODO: make this size configurable*/
```
29jm, line 62: Comment no longer needed I think

kernel/src/misc/wm/wm.c
```diff
@@ -36,7 +36,16 @@ static uint32_t id_count = 0;
...
+void buffer_dump(void* buff, size_t len) {
+    for (size_t i = 0; i < len; i++)
+    {
+        printf("%d:%#2X ", i, ((uint8_t*)buff)[i]);
+    }
+    printf("\n");
+}
+
 void init_wm() {
+    printk("wm window event buffer size: %d", WM_WINEVBUFF_SZ);
```
29jm: Remove debug prints

kernel/src/misc/wm/wm.c
```diff
@@ -155,6 +171,9 @@ void wm_get_event(uint32_t win_id, wm_event_t* event) {
 void wm_raise_window(wm_window_t* win) {
     list_t* win_iter;
     wm_window_t* win_bis;
+    wm_event_t active_ev;
+    wm_event_t win_ev;
+
```
29jm: Use only one event (call it `event`), and remove the extra blank line

kernel/src/misc/wm/wm.c
```diff
@@ -469,6 +490,9 @@ void wm_mouse_callback(mouse_t raw_curr) {
     // Move the cursor
     float dx = (raw_curr.x - raw_prev.x)*sens;
     float dy = (raw_curr.y - raw_prev.y)*sens;
+    // store the mouse event for inclusion in the ringbuffer
+    wm_event_t drag_ev;
+    wm_event_t uc_ev;
```
29jm, line 486: Don't declare those here, declare them in the innermost scope where they are used. Use unabbreviated names, `drag_event`, `move_event`, or just `event` (being declared close to usage gives enough context imo)

kernel/src/misc/wm/wm.c
```diff
@@ -554,15 +582,18 @@ void wm_mouse_callback(mouse_t raw_curr) {
 }
 
 void wm_kbd_callback(kbd_event_t event) {
+    wm_event_t kbd_ev;
```
29jm: Rename to `event` or `kbd_event` and assign directly its type etc... here. Reuse within the loop.

29jm (commented):
Great work! I can't review the ringbuffer changes right away (I'll get to them a bit later, maybe tomorrow), but I've reviewed the rest.

The terminal & paint work great, and things are much much better in doom, with much fewer stuck keys issues, though that still happens a bit. I'd consider the issue fixed though, this is good.

xadaemon:
> Great work! I can't review the ringbuffer changes right away (I'll get to them a bit later, maybe tomorrow), but I've reviewed the rest.
> 
> The terminal & paint work great, and things are much much better in doom, with much fewer stuck keys issues, though that still happens a bit. I'd consider the issue fixed though, this is good.

~on the doom bit I believe a larger queue size might help I went with a very conservative 5 but something like 20 might do the trick for doom~
edit: this doesn't make much sense since the buffer being overwritten shouldn't make it happen needs further investigation

kernel/src/misc/wm/wm.c
xadaemon: rather strange that clang_tidy didn't take care of this function ^^

29jm:
> ~on the doom bit I believe a larger queue size might help I went with a very conservative 5 but something like 20 might do the trick for doom~
> edit: this doesn't make much sense since the buffer being overwritten shouldn't make it happen needs further investigation

I think it could be that, after all there's only one keyup event for many more keydowns (from other keys being pressed at the same time, when moving with the arrows). This calls for another form of solution I think, extending ringbuffers feels a bit hacky (or perhaps in a case by case basis with a system call?). Outside the scope of this PR and issue anyway ^^

xadaemon:
yeah maybe some form of raw input might help

xadaemon:
I fixed every point you suggested (I think)

xadaemon:
> > ~on the doom bit I believe a larger queue size might help I went with a very conservative 5 but something like 20 might do the trick for doom~
> > edit: this doesn't make much sense since the buffer being overwritten shouldn't make it happen needs further investigation
> 
> I think it could be that, after all there's only one keyup event for many more keydowns (from other keys being pressed at the same time, when moving with the arrows). This calls for another form of solution I think, extending ringbuffers feels a bit hacky (or perhaps in a case by case basis with a system call?). Outside the scope of this PR and issue anyway ^^

should not be terribly hard to allow a program to request a buffer resize but like you said outside of this scope

29jm:
Looking at the ringbuffer implementation, there are things that bother me. It clearly appears to work, but I don't believe it is correct, in a few ways (do argue if I'm wrong of course). Three things:
+ `ringbuffer_available`:  [3,4,\_,\_,1,2], `w_base` after 4 (indice 2), `r_base` at 1 (indice 4), your implementation returns 2, when it should return 4. Something like this looks more correct (but still isn't, because of the next point):
    ```c
    if (ref->w_base < ref->r_base) {
        return ref->size - ref->r_base + ref->w_base;
    }

    return ref->w_base - ref->r_base;
    ```
+ same function: when the buffer is full, `w_base == r_base`, and your implementation returns 0. [Wikipedia](https://en.wikipedia.org/wiki/Circular_buffer#Circular_buffer_mechanics) highlights this issue: when dealing with two pointers (read and write), you need a way to distinguish empty and full buffers (I have a preference for dealing with `(r_base, data_size)` pairs myself for this reason, and ease of understanding)
+ reusing the example ringbuffer from above, consider writing 3 more elements to it, say 5, 6 and 7. The 1 gets overwritten (by the 7), yet your implementation keeps `r_base` at index 4, which means the next read will return the last element, and not the first, which is weird for a FIFO data structure. This is the reasoning behind the `if` I had added.

edit: if hell was a data structure it'd be a ringbuffer.

29jm:
Take a look at the last commit on `ringbuffers-evolved` (naming branches is easy but gives poor results ^^), the fix from your previous PR to working ringbuffers was really nothing much (fingers crossed it keeps working but this time I'm convinced it's mathematically correct).

xadaemon:
> Looking at the ringbuffer implementation, there are things that bother me. It clearly appears to work, but I don't believe it is correct, in a few ways (do argue if I'm wrong of course). Three things:
> 
>     * `ringbuffer_available`:  [3,4,_,_,1,2], `w_base` after 4 (indice 2), `r_base` at 1 (indice 4), your implementation returns 2, when it should return 4. Something like this looks more correct (but still isn't, because of the next point):
>       ```c
>       if (ref->w_base < ref->r_base) {
>           return ref->size - ref->r_base + ref->w_base;
>       }
>       
>       return ref->w_base - ref->r_base;
>       ```
> 
>     * same function: when the buffer is full, `w_base == r_base`, and your implementation returns 0. [Wikipedia](https://en.wikipedia.org/wiki/Circular_buffer#Circular_buffer_mechanics) highlights this issue: when dealing with two pointers (read and write), you need a way to distinguish empty and full buffers (I have a preference for dealing with `(r_base, data_size)` pairs myself for this reason, and ease of understanding)
> 
>     * reusing the example ringbuffer from above, consider writing 3 more elements to it, say 5, 6 and 7. The 1 gets overwritten (by the 7), yet your implementation keeps `r_base` at index 4, which means the next read will return the last element, and not the first, which is weird for a FIFO data structure. This is the reasoning behind the `if` I had added.
> 
> 
> edit: if hell was a data structure it'd be a ringbuffer.

I will try to address your concerns in order:

* in this case I made `available` to return the distance between the read and write pointers, as in how much space there is until data starts being overwritten;
* if it returns 0 it means there is no further space to add data without overwriting
* imho it adds unneeded complexity as much as it seems weird I don't think it's wrong behavior per se

> ringbuffers-evolved

**oh no they are evolving!** ^^

xadaemon:
I did some testing on the ringbuffer to rule em all it works! I think we can go with it yay ^^

xadaemon:
image !! merged your changes and will upload the commit

29jm:
Glorious!

xadaemon:
It's merged and uploaded if there are no other concerns I think we can merge this and end 2020 with one less issue ^^

29jm:
I'll try github's cli tool to merge & squash this correctly while still having the PR marked as merged, the online interface is of no help at all when you need more than just squashing everything...

xadaemon:
I did the merging already to my master branch if you just squash it should just work:tm:

29jm:
Aye, the cli tool doesn't do anything more than the web interface anyway. I really wish there was a way to make github cooperate better with git in this regard, but this'll be good enough ^^

29jm:
Thanks again for the help, have a great new year's eve!

xadaemon:
> Thanks again for the help, have a great new year's eve!

Have a great new year's eve too

PR #27, opened 2020-12-31
 remove a few things that somehow got back inside in the merge of the previous pr

xadaemon:
Sneaky buggers somehow got into the merge, that's pretty much all there is to this one, also added a comment to the debugging helper


29jm:
Good spot! Down that goes

PR #32, opened 2021-04-12
Made kernel size KB instead of KiB and added a global version string

PQCraft:
Made kernel size KB instead of KiB
Added a global version string
Changed some capitalization
Made multiboot error message more urgent


29jm:
Hi ! I love external contributions, but these are too superficial for me to merge, sorry :(

PQCraft:
F

PR #37, opened 2022-03-23
Implement basic tb hovering detection system

mikumikudice:
I'm trying to make it more usable by enabling the drag n' drop window system only when the mouse is hovering the title bar. It works, but when you shake the mouse, due to the fact that the mouse slides across the window, it's not too usable. I'm trying to fix the mouse mobility (``kernel/src/misc/wm/wm.c``) but I'm getting an internal page fault. The fallowing is my error output:
```
[paging.c] page fault caused by instruction at c01035fb from process 1:
[paging.c] the page at 0000002c wasn't present 
[paging.c] when a process tried to read from it
[paging.c] this process was in kernel mode
[stacktrace.c] stacktrace:
[stacktrace.c]  c0100fa0: paging_fault_handler
[stacktrace.c]  c0102200: isr_handler
[stacktrace.c]  c010aae3: isr48
[stacktrace.c]  c01039c0: wm_mouse_callback
[stacktrace.c]  c0108b60: mouse_handle_packet
[stacktrace.c]  c0108a10: mouse_handle_interrupt
[stacktrace.c]  c0101b40: irq_handler
[stacktrace.c]  c010a9ea: irq15
[stacktrace.c]  c0100000: .
[stacktrace.c]  c0100000: .
[stacktrace.c]  c0100000: .
```
I also updated the snow_draw_window to accept colors, so then we can use the function anywhere in various ways. Unfortunately it seems to not work properly when used in terminal.c, for example, resulting in a black window.

Any help in any of these two problems would be great.


mikumikudice:
I didn't opened an issue because my issue is related to my changes, not the repo itself

29jm:
I'm currently looking at it, will get back to you soon. Random notes:
- I looked at why windows are dragged at a speed different from mouse speed at some point, I think I it was because of rounding errors in the wm mouse callback, somewhere between dx/dy and window coordinates - will check again.
- you're introducing window decorations in the wm, which I think is inevitable to have something usable - I'm all for it. Initially I wanted to have the wm handle dumb rectangles, and to have everything UI in the userspace lib.

I really appreciate someone with design ability looking at this area btw

kernel/src/misc/wm/wm.c
```diff
@@ -489,65 +509,55 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
-    // Simple moves
-    if (!raw_prev.left_pressed && !raw_curr.left_pressed) {
+    } else {
         wm_window_t* under_cursor = wm_window_at(mouse.x, mouse.y);
-        wm_event_t event;
 
         if (under_cursor) {
-            rect_t r = rect_from_window(under_cursor);
+            wm_event_t event;
+            rect_t r = rect_from_window(win);
```
29jm: The crash is here, when this function is first called because of a mouse move: `win` is null. There are other branches in this function where `win` might be null that will cause problems.

kernel/src/misc/wm/wm.c
```diff
@@ -489,65 +509,55 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
+    if (lmb_is_down) {
+        // only drag window when hovering title bar and 
+        if ((dx || dy) && wm_hovering_tb_at(mouse.x, mouse.y)) {
+            rect_t rect = rect_from_window(win);
 
-            if (!(rect.left + dx < 0 || rect.right + dx >= (int32_t) fb.width ||
-                    rect.top + dy < 0 || rect.bottom + dy >= (int32_t) fb.height)) {
-                been_dragged = true;
+            win->x += dx;
+            win->y += dy;
```
29jm: The issue with the window moving slower than the cursor when dragging is related to these two lines:
- the cursor movements dx/dy are floats
- window coordinates are integers
- the decimal part of dx and dy are lost at every step of a drag motion

Two ways to fix this I can think of:
- make window coordinates floating points. This might be a source of headaches as it touches a lot of stuff.
- probably better, keep a cumulative dx/dy by having them static and incremented by the mouse movement each call. Also track the initial position of the dragged window at the start of the event, and recalculate its position with something like `initial_pos.x/y + cumulative_dx/dy`.

Actually I don't really like the first idea, second one is much easier and to the point.

modules/src/terminal.c
```diff
@@ -140,17 +148,18 @@ void redraw(str_t* text_buf, const str_t* input_buf) {
...
     // title bar
-    snow_draw_rect(win->fb, 0, 0, win->width, 20, 0x00222221);
-    snow_draw_border(win->fb, 0, 0, win->width, 20, 0x00000000);
-    snow_draw_string(win->fb, win->title, 4, 3, 0x00FFFFFF);
+    snow_draw_rect(win->fb, 0, 0, win->width, tb_height, clr.tb_color);
+    snow_draw_border(win->fb, 0, 0, win->width, tb_height, clr.tb_bcolor);
+    snow_draw_string(win->fb, win->title, tb_padding, tb_height / 3, clr.tb_tcolor);
     // border of the whole window
-    snow_draw_border(win->fb, 0, 0, win->width, win->height, 0x00555555);
+    snow_draw_border(win->fb, 0, 0, win->width, win->height, clr.w_bcolor);
```
29jm: Replacing all these drawing calls by `snow_draw_window(win, &clr);` works for me, gives a grey terminal as usual.

kernel/src/misc/wm/wm.c
29jm: using `under_cursor` makes it not crash.

29jm:
Left a few comments, let me know if you encounter new issues.

Do you think it would be possible to keep the structure of `wm_mouse_callback` more like it was, just adding one more condition to the "dragging" branch to catch the titlebar? I think it would also help with implementing the fix for the dragging speed issue.

Edit: this [page](https://github.com/29jm/SnowflakeOS/wiki/The-edit-debug-cycle) has info on how to debug stuff and might be of help.

kernel/src/misc/wm/wm.c
mikumikudice: thanks. I'll keep it up from here

mikumikudice:
I was trying to rewrite the ``wm_callback`` due to its slidy behaviour, but if it's better to you I can go back to the original code and just update it. Also, thanks by the debug stuff, it will be quite useful (and sorry by not finding it by myself).

29jm:
I made a draft version of the dragging fix fwiw, I just put it up on branch `dragging-improvs`. It works, with some not-smooth behavior when dragging a window close the the screen edges (due to the collision check with edges).

I'd be in favor of having `wm_callback` stay closer to before if you don't mind, I admit in good part because I know it and understand it well, and that function's job is a bit complicated.

mikumikudice:
I'm trying to implement now an highlight effect on the title bar (activated when hovered) but I don't know how to get the wm_window_t from a window_t. Is any function that give me back the wm version by the giving id?

29jm:
> Is any function that give me back the wm version by the giving id?

There are no ways to do that right now (no use case), what would that allow you to do if it was possible? edit: feel free to add one if it's clearly needed.

For the highlighting behavior of the titlebar, what do you think of the following way to approach this:
+ titlebar hovered is an event sent to the apps; like clicks. Say, an event when the mouse enters it, another when it goes out (details may get complicated?)
+ the window-drawing function in the userspace lib keeps track of that, adjusts color in the drawing loop (details on this are fuzzy for me right now sorry).

mikumikudice:
I was planning to simply add a field to the wm_window called ``is_hovered`` and use it to tell the snow_draw_window when use the highlight colors or not. But then I noticed that in terminal.c we only have access to the window_t, not the wm one. But the events sounds a good approach to me. I'll try to implement them

29jm:
Ah I understand. If you want to go for the polling route as opposed to event, a relatively easy way to do it (I think, but fuzzy) would be to add a command to `syscall_wm` that would return things like metadata about a window, hover for now but could be updated to include other stuff later.

mikumikudice:
I added two new syscalls (one to get the position of the window of a given id and one to get the hovering status) and implemented the highlight as well. Now I just want to enable paint to paint on click except when clicked on title bar, so if you could explain to me how ``paint.c`` works (and how to use it when running) it would be great

29jm:
Ok so `paint` uses widgets to render its UI and handle events, unlike say the terminal which as of now draws its UI manually. If you have experience with a library like GTK+, or even Qt to some extent, the UI library in Snowflake OS uses similar concepts. The idea at least in this project is, your app has a top level container, that you can put stuff in including other containers, and in the end what fills those containers will be widgets: labels, buttons, text fields, menu bars, custom widgets, etc.
As it stands, the top container is a vertical box (vbox_t), prepopulated with a titlebar_t widget. Whatever comes next in that container will be the app itself, `ui_set_root` is used to set that.

For paint the whole hierarchy goes like this:
| vbox_t: hidden root |                    |                        |                                  |
|---------------------|--------------------|------------------------|----------------------------------|
| ->                  | titlebar_t: paint  |                        |                                  |
| ->                  | vbox_t: root of UI |                        |                                  |
|                     | ->                 | hbox_t: menu bar       |                                  |
|                     |                    | ->                     | button widgets: colors and stuff |
|                     | ->                 | canvas_t: drawing area |                                  |

Hopefully after staring at it for a while this mess of a table will make sense :)

The containers are responsible for sizing their content (containers and widgets), and if needed the content also resizes its content etc all the way down the widget hierarchy. Widgets can specify how they like to be sized, e.g. capable of expanding horizontally, vertically, or not, etc. Containers also propagate UI events down to the widget they contain, e.g. you click on a container, container checks what widget was under your cursor, tells that widget (or container) that it got clicked at that spot (the coordinates of the click are converted to local coordinates from the widget's perspective (see the doc on coordinate functions in ui.c)).

How paint works is simply that it has a `canvas_t` widget that is setup to react to click and move events from the mouse, and because the widget "takes" space in the UI, it can just draw in that space (see canvas.c for details) when it is told to draw by its parent container (i.e. all the way up the chain, by paint.c in the main loop).

The previous limitation with the windows being draggable from everywhere meant that currently you have to click once in the canvas to start drawing, and click again to stop. With your changes to drag windows from titlebars, it should be doable now to draw in more conventional paint fashion, probably with some update to support it in the UI lib.

Take a look at "ui.h" to see how widgets are implemented (it's the usual kind of object oriented C à la gtk) and some tiny doc on it.

I hope this helps get you started on that side of things :)

mikumikudice:
I did understand how the paint work. I did this side correctly (or at least set everything to). Now I'm facing a problem with the mouse api. I tried to implement an "on release event" but I was not able to set wm.c to handle well dragging, once the click event is set once. My logic (for paint) was: "if clicked then start painting; if released stop to", but the release is set as soon as the click ends and I don't know how to really handle the physical button release. So, if you could give me some light (again) on this problem it would be great.

Also, I tried (and undone most of it) a lot of changes in the code, such as make wm_rect_t and rect_t the same thing (only removed the alias of wm_rect_t), fuse the button and color_button (paint and calc got really messed and bugged), but none of this worked because, sincerely, this part of the code is kinda messy and confusing, so I left this massive change (for example, reimplement the GUI/UI apis to fallow [this format](https://halt.software/dead-simple-layouts/)) for a new and later PR.

mikumikudice:
Actually, I think I found a way to do it. I just remembered that the dragging stops right when the mouse releases the button. I'll try to do something on this tomorrow

29jm:
Making drag events a thing in a clean way in the wm will take some effort, for sure. Having DRAG_START, DRAG_MOVE and DRAG_END events sounds useful. Let me know if doesn't want to work and I'll take a shot at it, compare notes.

By the way, I'm curious what problems you ran into merging button_t and color_button_t?  
One thing that I know poses problem is to try to extend an existing widget, e.g. `struct color_button_t { button_t base, <custom color_t members> };`, that just lead to a mess when I experimented with this.  
I wouldn't expect problems with using a single widget for a normal and a color (image or whatever) button though, as long as you provide different constructors or a constructor with proper parameters.

mikumikudice:
> Making drag events a thing in a clean way in the wm will take some effort, for sure. Having DRAG_START, DRAG_MOVE and DRAG_END events sounds useful. Let me know if doesn't want to work and I'll take a shot at it, compare notes.
> 
> By the way, I'm curious what problems you ran into merging button_t and color_button_t? One thing that I know poses problem is to try to extend an existing widget, e.g. `struct color_button_t { button_t base, <custom color_t members> };`, that just lead to a mess when I experimented with this. I wouldn't expect problems with using a single widget for a normal and a color (image or whatever) button though, as long as you provide different constructors or a constructor with proper parameters.

The buttons merge made the paint and calc windows turn black. The rect/wm_rect made the entire system go crazy and messed up. I think I can do it better, but when actively working for this as main goal.

mikumikudice:
Also, I just did it. Now it's just a matter of polishing. The click sometimes isn't detected, and the cursor flicks on the canvas. Put that aside, everything works as desired

29jm:
Hoo it's awesome to see it work that way! Great job. I'll give it a proper review tomorrow.

kernel/include/kernel/wm.h
```diff
@@ -19,29 +31,41 @@
...
+void wm_assign_position(wm_window_t* win);
+void wm_assign_z_orders();
+void wm_raise_window(wm_window_t* win);
+list_t* wm_get_window(uint32_t id);
+void wm_print_windows();
+list_t* wm_get_windows_above(wm_window_t* win);
+wm_rect_t wm_mouse_to_rect(mouse_t mouse);
+void wm_draw_mouse(wm_rect_t new);
+void wm_mouse_callback(mouse_t curr);
+void wm_kbd_callback(kbd_event_t event);
```
29jm: These functions are not meant to be used outside of wm.c (in fact they could be static), they shouldn't be declared here. If a prototype is needed it should be added to the top of wm.c, just before the first function definition.

kernel/include/kernel/wm.h
```diff
@@ -19,29 +31,41 @@
 typedef struct _wm_window_t {
     fb_t ufb;
     fb_t kfb;
-    int32_t x;
-    int32_t y;
+    point_t* pos;
     uint32_t id;
     uint32_t flags;
     ringbuffer_t* events;
+    bool is_hovered;
```
29jm: I'm not sure I like this flag, as "being hovered" it's not a property of the window, the window doesn't choose, it's decided by the wm.

I think the answer to "being hovered" should be a function call, you already implemented one I think. It costs basically nothing to recompute on demand, and it makes sure that the property is always up to date.

kernel/include/kernel/wm.h
```diff
@@ -9,8 +11,18 @@
 
 #include <kernel/uapi/uapi_wm.h>
 
+#include <gui_meta.h>
+
 #define WM_NOT_DRAWN  ((uint32_t) 1 << 31) // Window has _never_ been called wm_render_window
 
+typedef struct {
+    int32_t x, y, w, h;
+} rect_t;
```
29jm: I don't understand that, given that we have `wm_rect_t`. We shouldn't have two types for the same thing.

ui/include/ui.h
```diff
@@ -3,6 +3,7 @@
 #include <snow.h>
 
 #include <kernel/uapi/uapi_wm.h> // For fb_t
+#include <kernel/wm.h>
```
29jm: This isn't right. The userspace-exposed APIs are in "uapi" headers, for User API. When developing stuff here, you should assume that userspace cannot access "kernel/wm.h" at all.

The idea of uapi headers is to provide a limited interface to the kernel, without having to duplicate declarations because they are also included by the actual kernel headers (i.e. wm.h).

Makefile
```diff
@@ -33,12 +33,12 @@ endif
...
-# AR=ar
-# AS=as
-# CFLAGS+=-target i386-pc-none-eabi -m32
-# CFLAGS+=-mno-mmx -mno-sse -mno-sse2
+#CC=clang
+#LD=ld
+#AR=ar
+#AS=as
+#CFLAGS+=-target i386-pc-none-eabi -m32
+#CFLAGS+=-mno-mmx -mno-sse -mno-sse2
```
29jm: Please remove the change.

kernel/include/kernel/wm.h
29jm: Ok so I understood, seeing the change in "ui.h". These structures here (thinking of `point_t` below too) that you want to access in both kernel and userspace belong in "uapi_wm.h", please move them.

29jm: Also I see this is used only in the userspace UI library, so I vote for this definition to move back to "ui.h".

kernel/include/kernel/wm.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <kernel/fb.h>
+#include <kernel/mouse.h>
+#include <kernel/kbd.h>
```
29jm: These should be removed at the same time as the prototypes below.

kernel/src/misc/wm/wm.c
```diff
@@ -13,21 +13,6 @@
...
-void wm_assign_z_orders();
-void wm_raise_window(wm_window_t* win);
-list_t* wm_get_window(uint32_t id);
-void wm_print_windows();
-list_t* wm_get_windows_above(wm_window_t* win);
-rect_t wm_mouse_to_rect(mouse_t mouse);
-void wm_draw_mouse(rect_t new);
-void wm_mouse_callback(mouse_t curr);
-void wm_kbd_callback(kbd_event_t event);
-
```
29jm: Please reinstate. Think of it as encapsulation, module by module, exposing only what's needed.

kernel/src/misc/wm/wm.c
```diff
@@ -447,7 +433,25 @@ wm_window_t* wm_window_at(int32_t x, int32_t y) {
     return NULL;
 }
 
-void wm_draw_mouse(rect_t new) {
+/* Returns the foresmost window which title bar is being
+ * hovered by the mouse (x, y), NULL if none match
+ */
+bool wm_hovering_tb_at(int32_t x, int32_t y) {
```
29jm: The doc comment is out of date looks like, but I think the API it suggests might be better: returning the window whose titlebar is being hovered, if any.

kernel/src/misc/wm/wm.c
```diff
@@ -469,8 +473,8 @@ void wm_draw_mouse(rect_t new) {
  */
 void wm_mouse_callback(mouse_t raw_curr) {
     static mouse_t raw_prev;
-    static wm_window_t* dragged = NULL;
-    static bool been_dragged = false;
+    static wm_window_t* dragged = NULL,* hovered = NULL;
+    static uint16_t been_dragged = false;
```
29jm: Get out of here with this `uint16_t` heresy x)  
We are a `bool` shop here.

kernel/src/misc/wm/wm.c
```diff
@@ -489,29 +493,45 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
+    hovered = wm_window_at(mouse.x, mouse.y);
+
+    if(hovered){
+        hovered->is_hovered = wm_hovering_tb_at(mouse.x, mouse.y);
+    }
+
+    // Click, raises and drag starts
     if (!raw_prev.left_pressed && raw_curr.left_pressed) {
-        wm_window_t* clicked = wm_window_at(mouse.x, mouse.y);
+        dragged = wm_window_at(mouse.x, mouse.y);
```
29jm: I think you can get away with `dragged = hovered` here.

kernel/src/misc/wm/wm.c
```diff
@@ -489,29 +493,45 @@ void wm_mouse_callback(mouse_t raw_curr) {
     mouse.x = mouse.x < 0 ? 0 : mouse.x;
     mouse.y = mouse.y < 0 ? 0 : mouse.y;
 
-    // Raises and drag starts
+    hovered = wm_window_at(mouse.x, mouse.y);
+
+    if(hovered){
+        hovered->is_hovered = wm_hovering_tb_at(mouse.x, mouse.y);
+    }
+
```
29jm: As with `is_hovered`, this should go away.

From what I can tell, this is used later in this function and in the wm syscall, in both cases it can just be recomputed on the fly. This means the function to compute it will have its prototype in "wm.h", so it is exposed to other modules.

kernel/src/misc/wm/wm.c
```diff
@@ -489,29 +493,45 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
-            rect_t rect = rect_from_window(dragged);
+            wm_rect_t rect = rect_from_window(dragged);
 
             if (!(rect.left + dx < 0 || rect.right + dx >= (int32_t) fb.width ||
                     rect.top + dy < 0 || rect.bottom + dy >= (int32_t) fb.height)) {
-                been_dragged = true;
 
-                dragged->x += dx;
-                dragged->y += dy;
+                if(dragged->is_hovered){
```
29jm: Use a space after `if` and between the condition and body braces, `if (...) {`. See [this](https://github.com/29jm/SnowflakeOS/blob/master/CONTRIBUTING.md) for other arbitrary rules.

kernel/src/misc/wm/wm.c
```diff
@@ -526,28 +546,26 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
         }
-
-        been_dragged = false;
         dragged = NULL;
     }
 
-    // Simple moves
-    if (!raw_prev.left_pressed && !raw_curr.left_pressed) {
+    // The mouse's simply moving
+    if (raw_prev.x != raw_curr.x || raw_prev.y != raw_curr.y) {
```
29jm: Curious if you had issues with using `dx || dy` as the condition here. If you did, it might good to do that in the other condition that uses `dx || dy`. Otherwise `dx || dy` looks kind of nicer, no?

kernel/src/sys/proc.c
```diff
@@ -44,7 +44,7 @@ process_t* proc_run_code(uint8_t* code, uint32_t size, char** argv) {
         argv++;
     }
 
-    // Allocate one page more than the program size to accomodate static
+    // Allocate one page more than the program size to accommodate static
     // variables.
```
29jm: Could you remove this comment entirely? It's out of date, now the exact number of pages needed is allocated, relying on progbits behavior to store the actual zeros inside executable files, as the TODO comment below indicates.

Incredibly well spotted otherwise.

kernel/src/sys/syscall.c
```diff
@@ -139,6 +139,26 @@ static void syscall_wm(registers_t* regs) {
...
+        case WM_CMD_GET_POS: {
+            list_t* item = wm_get_window(regs->ecx);
+            if(item){
+                regs->eax = list_entry(item, wm_window_t)->pos;
+            } else {
+                printke("the given window id (%d) is unknown", regs->ecx);
+                regs->eax = -1;
+                break;    
+            }
+        } break;
```
29jm: It seems to me like this is unused. You use it in `snow_open_window` but it will get out of date on the first drag event.

Do you see a need for it in the near future?

kernel/src/sys/syscall.c
```diff
@@ -139,6 +139,26 @@ static void syscall_wm(registers_t* regs) {
...
+        case WM_CMD_IS_HOVERED: {
+            list_t* item = wm_get_window(regs->ecx);
+            if(item){
+                regs->eax = (list_entry(item, wm_window_t))->is_hovered;
+            } else {
+                printke("the given window id (%d) is unknown", regs->ecx);
+                regs->eax = -1;
+                break;    
+            }
+        } break;
```
29jm: Should be renamed to `WM_CMD_IS_DRAGGED` as it is more accurate. Hovered is usually used for when the cursor is above stuff without clicking it.

libc/include/abort.h
```diff
@@ -0,0 +1,10 @@
+#pragma once
+#include <stdio.h>
+
+#ifndef _KERNEL_
+
+#include <stdlib.h>
+
+#endif
+
+void abort();
```
29jm: I don't see the need for this, the prototype for `abort` is already in "stdlib.h".

libc/src/stdlib/abort.c
```diff
@@ -1,10 +1,4 @@
-#include <stdio.h>
-
-#ifndef _KERNEL_
-
-#include <stdlib.h>
-
-#endif
+#include <abort.h>
```
29jm: Should be reverted.

modules/src/paint.c
```diff
@@ -91,6 +92,17 @@ int main(int argc, char* argv[]) {
...
+            canvas->widget.bounds);
+
+        // start drawing
+        if (event.type == WM_EVENT_CLICK && under_canvas) {
+            canvas->is_drawing = true;
+        // stop to
+        } else if(event.type == WM_EVENT_RELEASE || !under_canvas) {
+            canvas->is_drawing = false;
+        }
+
```
29jm: This should be done in "canvas.c". The appropriate signals should be routed to the widget the same way `WM_EVENT_MOUSE_MOVE` is for instance (see `ui_handle_input` in "ui.c" and functions like `lbox_on_click` in "lbox.c" which is the main container type).

29jm: Feel free to do this at a later stage, as it's kind of a boiler plate boring style change. It leads to clean APIs though!

snow/include/gui_meta.h
```diff
@@ -0,0 +1,13 @@
...
+*/
+#define tb_height 30
+#define tb_padding 8
+
+typedef struct {
+    int64_t bg_color, // background color
+    base_color, border_color, text_color, highlight, // title bar colors
+    border_color2; // window border color
+    bool is_hovered; // enables tb highlight
+} color_scheme_t;
```
29jm: I'm thinking this header file can be eliminated:
+ Put the padding defines in "uapi_wm.h" (rename them to scarier looking identifiers if possible, WM_TITLEBAR_.... or something)
+ `color_scheme_t` can be declared in "ui.h"

And unrelatedly `is_hovered` in `color_scheme_t` looks out of place for the same reason it did in `wm_window_t`, it's not logically part of the state of the color scheme. It's the state of the parent window (ironically, in the wm it wasn't, but uh, different contexts and info available).

edit: actually it's still not the state of the parent window, it's the state of whatever the HOVERED syscall returns.

snow/include/snow.h
```diff
@@ -6,6 +6,9 @@
 #include <kernel/uapi/uapi_syscall.h>
 #include <kernel/uapi/uapi_wm.h>
 #include <kernel/uapi/uapi_kbd.h>
+#include <kernel/wm.h>
```
29jm: To be removed.

snow/include/snow.h
```diff
@@ -24,6 +27,8 @@ typedef struct {
     fb_t fb;
     uint32_t id;
     uint32_t flags;
+    point_t* pos;
+    bool is_hovered;
```
29jm: `pos` should not be here (it'll get out of sync. In the wm it did not, because it was the source of truth for that info) and `is_hovered` also.

snow/src/gui.c
```diff
@@ -46,15 +57,26 @@ void snow_close_window(window_t* win) {
...
+    clr->is_hovered = syscall2(SYS_WM, WM_CMD_IS_HOVERED, win->id);
+
     // background
-    snow_draw_rect(win->fb, 0, 0, win->width, win->height, 0x00AAAAAA);
+    snow_draw_rect(win->fb, 0, 0, win->width, win->height, clr->bg_color);
     // title bar
-    snow_draw_rect(win->fb, 0, 0, win->width, 20, 0x00222221);
-    snow_draw_border(win->fb, 0, 0, win->width, 20, 0x00000000);
-    snow_draw_string(win->fb, win->title, 4, 3, 0x00FFFFFF);
+    if(clr->is_hovered){
```
29jm: Formatting + get rid of intermediate variable `is_hovered`.

ui/include/ui.h
```diff
@@ -118,6 +114,8 @@ typedef struct {
 } pixel_buffer_t;
 
 bool point_in_rect(point_t p, rect_t r);
+rect_t wm_rect_to_rect(wm_rect_t r);
+point_t rect_to_point(wm_rect_t r);
```
29jm: `rect_to_rect` is weird and I think a result of the confusion around uapi_wm stuff.

ui/src/button.c
```diff
@@ -3,36 +3,59 @@
...
-    (void) p;
-    // TODO: show the clicked state graphically
+    rect_t r = ui_get_absolute_bounds((widget_t*) button);
 
-    if (button->on_click) {
-        button->on_click(button);
-    }
+    if (point_in_rect(p, r)) {
+        button->is_clicked = true;
+    } else button->is_clicked = false;
```
29jm: `point_in_rect` is not useful here. The idea with callbacks here is that they are called when relevant, meaning `p` is always inside the widget's bounds here, no need to check. `is_clicked` can just be set to true, which leads to the question, when is it set to false? -> missing a callback for RELEASE (see the comment on paint.c about adding it).

ui/src/ui.c
```diff
@@ -13,6 +13,19 @@ bool point_in_rect(point_t p, rect_t r) {
...
 }
 
+rect_t wm_rect_to_rect(wm_rect_t r) {
+    return (rect_t) {
+        .x = r.left,
+        .y = r.top,
+        .w = r.right - r.left,
+        .h = r.bottom = r.top
+    };
+}
```
29jm: Remove.

ui/src/button.c
```diff
@@ -3,36 +3,59 @@
...
 }
 
 void button_on_free(button_t* button) {
     free(button->text);
 }
 
-button_t* button_new(char* text) {
+button_t* button_new(char* text, color_scheme_t* clr) {
```
29jm: I have an idea to simplify all the new `clr` parameters that I think is kind of cool:
+ Have `color_scheme` be a field of `widget_t`
+ Set it to a default predefined color scheme for the root widget in "ui.c", allow it to be changed by a function like `ui_set_color_scheme`
+ Leave it to null everywhere else (it'll be null by default)
+ Make a function that takes a widget, goes up the hierarchy until it finds one with a non-null color scheme (the root, but not necessarily)
+ Use the output of that function whenever you need a style in a specific widget implementation, i.e. in their draw callbacks. No need to cache results for now.

Now the color scheme is global to an application's UI, and you can even change it on a per-container or per-widget basis, with no API overhead.

29jm (commented):
Some refactor to do, nothing major I think but little things add up over a large amount of changes.

I left an improvement idea for color schemes in button.c that I meant to put here, but heh :)

I noticed that clicks in the calc app and no longer worked. It's probably a minor thing, feel free to address review comments first and we'll look at it later.

kernel/src/misc/wm/wm.c
mikumikudice: Thy may was a faulty act. I'm not used to bool type in C (even in my own sources I end up using u16 instead of bool sometimes). No reason at all for it. I may have happened because I removed then re-added the variable

kernel/src/misc/wm/wm.c
mikumikudice: I didn't left some time to set up the clang compiler to Autoformat the code for me. But I'm aware of the code style, sorry. My fault

kernel/src/misc/wm/wm.c
mikumikudice: I have some ADD so I tend to left sometimes things out of mind. I was my first real PR and I'm still getting used to git versioning. I think if I had compared the versions I would have used this better approach. I simply completely forgot that I already had the delta position lol I'll not let this happen again. My bad

libc/src/stdlib/abort.c
mikumikudice: Got it. I've added for debug purposes and ended let it left behind. Tomorrow I'll make a last commit fixing all of this

ui/src/button.c
mikumikudice: That's an excellent idea. That's why you are in charge here, sir

kernel/src/sys/syscall.c
mikumikudice: It actually is used once when setting up the other window type. But actually, this can be removed once the two windows are merged or something like that. Also, the position may be not that necessary for the applications themselves, so if you don't think both windows should share the same position, I can remove it

kernel/src/sys/proc.c
mikumikudice: Since I always spell things wrong (and it gets worse due the fact English is not my native language) I installed a VS Code extension that checks grammar for me

kernel/src/misc/wm/wm.c
29jm: Don't sweat it ! It's totally normal that this kind of thing gets caught in review - it's the entire purpose of the review, there is no mistake on your part here.

Of course during development you focus on getting things to work, reviews are for polishing stuff. Same with all my other comments here, it's really no big deal. The underlying change is clearly good, it's impressive that it's your first PR.

kernel/src/sys/syscall.c
29jm: I don't think `wm_window_t` and `window_t` will ever be merged, as they are really part of different applications: the window manager in the kernel (hopefully at some point moved to user space), and the UI library which defines its own concept of window.

I think it makes sense that a gui app can query the position of its window, so please keep it.  
However having the position stored in the UI-defined window type (see comment in "snow.h") doesn't make sense, because then the UI lib has to keep it in sync all the time (user moves window -> DRAGGED event callback -> syscall to get new pos) for no good reason the majority of the time (an app that needs to know its window position can make the call itself on demand).

mikumikudice: Woops, it seems that in my "freak go back" commit I accidentally removed the part where I set the ``point_t* window_t::pos`` to the syscall return. My fault again. In this case, I believe, we wouldn't need to update the position every event because it's a pointer to the actual structure, but if you wish so, I'm getting rid of it.

mikumikudice:
Despite paint now highlights the buttons on strange colors (I'm not sure why once I'm just adding 0x030303 to them) everything keeps working as should (or better) and all changes requested have been effectuated

mikumikudice:
No, wait. Dragging stopped to work properly. Quick fix soon.
Edit: done! Ready to merge

kernel/src/sys/syscall.c
29jm: That would not work, if it pointed to kernel memory, accessing it would crash the app (let me know if it doesn't, that would be a fundamental bug). The reverse could work, using process-level memory in the kernel and process (it's used to share window buffers, `ufb` in `wm_window_t`), but we don't want to complicate things that much.

29jm:
Thanks for addressing all that! I will bother you tomorrow for a few more small things (sorry ^^'), I will have time to test it better also, but this is shaping up, very cool.

mikumikudice:
No problem, I'm here to help

kernel/src/misc/wm/wm.c
```diff
@@ -10,21 +10,23 @@
 #include <list.h>
 #include <stdlib.h>
 
+#include <snow.h>
+
```
29jm: This isn't right, snow.h has no place in the kernel. The idea of libsnow is to provide a C interface to the kernel's system calls that don't fit the C std lib. Kernel structures and defines from the WM you want shared with userspace go into "uapi_wm.h".

kernel/src/misc/wm/wm.c
```diff
@@ -447,7 +450,20 @@ wm_window_t* wm_window_at(int32_t x, int32_t y) {
...
+ */
+bool wm_is_window_beeing_hovered(wm_window_t* win, int32_t x, int32_t y) {
+    wm_rect_t r = rect_from_window(win);
+
+    if (y >= r.top && y <= r.top + UI_TB_HEIGHT && x >= r.left && x <= r.right) {
+        return true;
+    } else {
+        return false;
+    }
+}
```
29jm: + doc comment should read "Returns whether the cursor is above the given window's titlebar."
+ change the name to `wm_is_titlebar_hovered`
+ remove parameters x, y
+ use `mouse.x`, `mouse.y` in place of x, y

kernel/src/misc/wm/wm.c
```diff
@@ -481,6 +496,7 @@ void wm_mouse_callback(mouse_t raw_curr) {
     float dx = (raw_curr.x - raw_prev.x)*sens;
     float dy = (raw_curr.y - raw_prev.y)*sens;
 
+
```
29jm: Remove extra whitespace.

kernel/src/misc/wm/wm.c
```diff
@@ -489,32 +505,47 @@ void wm_mouse_callback(mouse_t raw_curr) {
...
 
     // Drags
     if (raw_prev.left_pressed && raw_curr.left_pressed) {
-        if (dragged && (dx || dy)) {
-            rect_t rect = rect_from_window(dragged);
+        bool hovered = false;
+
+        if(dragging){
+            hovered =
+            wm_is_window_beeing_hovered(dragging, mouse.x, mouse.y);
```
29jm: Formatting (if + remove line break)

kernel/src/sys/proc.c
```diff
@@ -44,9 +44,6 @@ process_t* proc_run_code(uint8_t* code, uint32_t size, char** argv) {
         argv++;
     }
 
-    // Allocate one page more than the program size to accomodate static
-    // variables.
-    // TODO: this assumes .bss sections are marked as progbits
```
29jm, line 49: Only the first two lines are to be removed, the TODO line is up to date and actually important :)

kernel/src/sys/syscall.c
```diff
@@ -139,6 +139,16 @@ static void syscall_wm(registers_t* regs) {
                 wm_param_event_t* param = (wm_param_event_t*) regs->ecx;
                 wm_get_event(param->win_id, param->event);
         } break;
+        case WM_CMD_IS_DRAGGED: {
+            list_t* item = wm_get_window(regs->ecx);
+            if (item) {
+                regs->eax = (list_entry(item, wm_window_t));
```
29jm: There is no way this works, a pointer to kernel memory will page fault when accessed from userspace. From its use in "gui.c", you treat the pointer as a `bool`: that won't crash but that also does not make sense, a valid pointer will always evaluate to true.

Use `regs->eax = wm_titlebar_is_hovered(...)` after putting the prototype of `wm_titlebar_is_hovered` in "wm.h".

libc/src/stdlib/abort.c
```diff
@@ -1,11 +1,3 @@
-#include <stdio.h>
-
-#ifndef _KERNEL_
-
-#include <stdlib.h>
-
-#endif
-
```
29jm: Revert "abort.c" to how it was.

modules/src/paint.c
```diff
@@ -1,5 +1,6 @@
 #include <snow.h>
 #include <ui.h>
+#include <kernel/wm.h>
```
29jm: Don't include anything from the WM here, not even "uapi_wm.h" (again "wm.h" is out of reach from userspace by design).

The "snow.h" include already does what you want.

snow/include/snow.h
```diff
@@ -17,13 +17,32 @@ int32_t syscall3(uint32_t eax, uint32_t ebx, uint32_t ecx, uint32_t edx);
...
+    border_color2; // window border color
+} color_scheme_t;
+
+#define UI_DEFAULT_COLOR &(color_scheme_t){\
+    .base_color = 0x757575,\
+    .border_color = 0x000000,\
+    .border_color2 = 0x121212,\
+    .text_color = 0xFFFFFF,\
+    .highlight = 0x030303,\
+}
```
29jm: Declare this as
```c
static default_color_scheme = (color_scheme_t) {
    ...
}
```
in "ui.c", and remove the uses of "UI_DEFAULT_COLOR".

In `ui_app_new` in "ui.c", set the root widget's color scheme to `&default_color_scheme`.

snow/include/snow.h
```diff
@@ -17,13 +17,32 @@ int32_t syscall3(uint32_t eax, uint32_t ebx, uint32_t ecx, uint32_t edx);
...
+
 typedef struct {
     char* title;
     uint32_t width;
     uint32_t height;
     fb_t fb;
     uint32_t id;
     uint32_t flags;
+    color_scheme_t* color;
```
29jm: Don't add this here:
+ the color scheme is a GUI concept
+ `window_t` here is just that: a window. Think of when you open a full screen game (e.g. doom) - the color scheme concept would make no sense there right? this is why it's out of place.

snow/src/gui.c
```diff
@@ -47,14 +48,28 @@ void snow_close_window(window_t* win) {
  * the screen.
  */
 void snow_draw_window(window_t* win) {
+    const color_scheme_t* clr = win->color;
```
29jm: Hardcode a color scheme here. `snow_draw_window` is only used by non-GUI lib apps.

I know this leads to some duplication, between the default UI colors and this. I prefer this to breaking the separation between the "dumb window" vs. widget lib.

The cause is a probable design flaw here, which is that the WM should be the one drawing titlebars and borders, which would unify all this and likely make more sense. A change for another time though :)

snow/src/gui.c
```diff
@@ -47,14 +48,28 @@ void snow_close_window(window_t* win) {
...
+    if (is_hovered) {
+        int64_t hl = clr->base_color + clr->highlight;
+
+        // color overflow
+        if (hl < clr->base_color) hl = 0xffffff;
+
+        snow_draw_rect(win->fb, 0, 0, win->width, UI_TB_HEIGHT, clr->base_color + clr->highlight);
+    } else {
+        snow_draw_rect(win->fb, 0, 0, win->width, UI_TB_HEIGHT, clr->base_color);
+    }
```
29jm: Oh no! colors are overflowing :D

- Could you make use of `uint32_t`s for all things color-related? This is the logical type to use, unless negative colors are a thing.
- Use braces for the inner `if` + formatting.

Fun fact, 64-bit arithmetic will be handled by the functions at `arith64.c`, SnowflakeOS being 32-bit.

ui/include/ui.h
```diff
@@ -128,22 +131,25 @@ rect_t ui_get_absolute_bounds(widget_t* widget);
...
 void button_set_on_click(button_t* button, void (*callback)(button_t*));
 void button_set_text(button_t* button, const char* text);
+void color_button_on_resize(color_button_t* button);
+
+color_scheme_t* color_to_scheme(uint32_t color);
 
 titlebar_t* titlebar_new(const char* title, const uint8_t* icon);
 void titlebar_set_title(titlebar_t* tb, const char* title);
 
-color_button_t* color_button_new(uint32_t color, uint32_t* to_set);
```
29jm: Please reinstate this one.

ui/src/button.c
```diff
@@ -3,40 +3,55 @@
...
-        button->on_click(button);
-    }
+    //rect_t r = ui_get_absolute_bounds((widget_t*) button);
+    button->is_clicked = true;
+}
+
+void button_on_release(button_t* button, point_t p) {
+    (void) p;
+
+    //rect_t r = ui_get_absolute_bounds((widget_t*) button);
```
29jm: Remove commented out code.

ui/include/ui.h
```diff
@@ -128,22 +131,25 @@ rect_t ui_get_absolute_bounds(widget_t* widget);
...
+
 hbox_t* hbox_new();
 void hbox_add(hbox_t* hbox, widget_t* widget);
 
 vbox_t* vbox_new();
 void vbox_add(vbox_t* vbox, widget_t* widget);
 void vbox_clear(vbox_t* vbox);
 
-button_t* button_new(char* text);
+button_t* button_new(char* text, color_scheme_t* clr);
```
29jm: Remove `clr` parameter.

ui/src/canvas.c
```diff
@@ -1,25 +1,34 @@
 #include <ui.h>
-
 #include <stdlib.h>
+#include <stdio.h>
```
29jm: Remove the debug stdio include.

ui/src/canvas.c
```diff
@@ -1,25 +1,34 @@
...
+    canvas->is_drawing = true;
+
+    canvas->last_pos = (point_t) { bounds.x + p.x, bounds.y + p.y };
+    canvas->new_pos = canvas->last_pos;
 }
 
 void canvas_on_mouse_move(canvas_t* canvas, point_t p) {
     rect_t bounds = ui_get_absolute_bounds((widget_t*) canvas);
 
+    if (!point_in_rect(p, bounds)) canvas->is_drawing = false;
```
29jm: + Formatting
+ Have you seen this condition entered? If it happens I'll see to fix it, as it shouldn't.

ui/src/canvas.c
```diff
@@ -1,25 +1,34 @@
...
+
     canvas->needs_drawing = true;
     canvas->last_pos = canvas->new_pos;
     canvas->new_pos = (point_t) { p.x+bounds.x, p.y+bounds.y };
 }
 
+void canvas_on_mouse_release(canvas_t* canvas, point_t p) {
+    rect_t bounds = ui_get_absolute_bounds((widget_t*) canvas);
+
+    // start drawing
```
29jm: That comment is drunk, remove it ^^

ui/src/canvas.c
```diff
@@ -39,9 +48,11 @@ canvas_t* canvas_new() {
     canvas_t* canvas = zalloc(sizeof(canvas_t));
 
     canvas->widget.flags = UI_EXPAND;
+    canvas->widget.color = NULL;
```
29jm: `zalloc` clears the allocated memory to zero, setting color to `NULL` is redundant here.

ui/src/lbox.c
```diff
@@ -143,8 +157,11 @@ lbox_t* lbox_new(uint32_t direction) {
     lbox->widget.on_draw = (widget_draw_t) lbox_on_draw;
     lbox->widget.on_resize = (widget_resize_t) lbox_on_resize;
     lbox->widget.on_mouse_move = (widget_mouse_moved_t) lbox_on_mouse_move;
+    lbox->widget.on_mouse_release = (widget_mouse_release_t) lbox_on_mouse_release;
     lbox->direction = direction;
 
+    lbox->widget.color = UI_DEFAULT_COLOR;
```
29jm: Remove.

ui/src/ui.c
```diff
@@ -13,6 +13,17 @@ bool point_in_rect(point_t p, rect_t r) {
...
+point_t rect_to_point(wm_rect_t r) {
+    return (point_t) {.x = r.left, .y = r.top};
+}
+
+/* Returns the color scheme of the widget or of the closest parent
+ */
+color_scheme_t* get_widget_color(widget_t* widget) {
+    if (widget->color != NULL) return widget->color;
+    else return get_widget_color(widget->parent);
+}
```
29jm: + Add a first `if (!widget) { return &default_color_scheme; } just for safety
+ Formatting :)

ui/src/ui.c
```diff
@@ -129,4 +150,18 @@ point_t ui_absolute_to_local(widget_t* widget, point_t point) {
...
+    *clr = (color_scheme_t){
+        .base_color = color,
+        .bg_color = color / 3,
+        .border_color = color / 4,
+        .text_color = color - color / 2,
+        .highlight = color + 0x030303,
+    };
+
+    return clr;
 }
```
29jm: Do you think this could be removed, as it is unused?

If it is to stay please:
+ add a doc comment
+ return a `color_scheme_t` by value, not `zalloc`'ed
+ space before `{`

29jm (commented):
Some more comments as promised.  
One more, could you fix the warnings when compiling the project? `make clean ; make strict`

Issues I noticed that I'll try to debug:
+ can't paint in the top part of the canvas in the paint app - UI warnings in the terminal, likely related.
    + the extra `if` check in "canvas.c" mentioned in review is responsible for the drawing bug at the top of the canvas.
    + the UI warnings are due to the buttons requesting a bigger height than the height of the menu container which is set to 20px.
        + this can be fixed multiple ways: increase menu height, or set buttons to expand vertically, ... there is room here for an API design improvement, though I haven't found it yet.
        + these warnings aren't super important, the container fixes the bounds anyway.
+ "Clear" button in paint doesn't work, buttons in calc app in general have no effect.
     + ok this one was straigtforward, please re-add the call to the clicked callback.

kernel/src/sys/syscall.c
29jm: you'll also need to declare `wm_get_window` in "wm.h" to access it here in "syscall.c" (without warnings I mean)

ui/src/canvas.c
29jm: edit: delete `if`, see main review comment.

kernel/src/sys/syscall.c
mikumikudice: A null pointer is returned from ``wm_get_window`` if none window is found, that is, what I was testing, as NULL is treated as false in C. But in any way, I was kinda crazy when I wrote this code. I think I refactored it without noticing. Or maybe I had an idea and ended up with another, without conciliating them. I'll make this a boolean syscall, for sure.

snow/include/snow.h
mikumikudice: Where should I set root as so? I thought on that but I have understood that the derivative files that implemented the UI "constructors"

snow/src/gui.c
mikumikudice: got it

snow/src/gui.c
mikumikudice: This was just a "make sure" for the highlights I've implemented on the paint color buttons. I'm still not sure why they get strange colors. By the way it have been fixed

ui/src/ui.c
mikumikudice: I was already considering remove this. It's residual code from my attempt to merge (and understand) the buttons types

mikumikudice:
Everything is set up and up to date. Thanks for lighting up for me some of my faulty acts. Now, for sure I guess, we're ready to merge

kernel/src/misc/wm/wm.c
```diff
@@ -447,7 +448,21 @@ wm_window_t* wm_window_at(int32_t x, int32_t y) {
     return NULL;
 }
 
-void wm_draw_mouse(rect_t new) {
+/* Returns true if the given window's title bar is being
+ * hovered by the mouse (x, y), false otherwise
+ */
+bool wm_is_window_beeing_hovered(wm_window_t* win) {
```
29jm: typo beeing vs. being, remove (x, y)

kernel/src/misc/wm/wm.c
```diff
@@ -215,18 +215,19 @@ void wm_raise_window(wm_window_t* win) {
  * TODO: revamp entirely.
  */
 void wm_assign_position(wm_window_t* win) {
+    win->pos = (point_t*) kmalloc(sizeof(point_t));
```
29jm: sorry I didn't catch that earlier, the window position shouldn't be a pointer.

Rationale is simply that memory allocation is way more complicated (see malloc.c) than stack, and pointers themselves are known to cause complications (see "returning kernel pointer from syscall" from previous review), and all this is in the kernel (admittedly dumb and my fault for the WM), which is not a place you want to complicate without good reason.

kernel/src/sys/syscall.c
```diff
@@ -138,7 +138,19 @@ static void syscall_wm(registers_t* regs) {
...
+
+            if (item != NULL) {
+                wm_window_t* win = list_entry(item, wm_window_t);
+                return win->being_dragged;
+            } else {
+                printke("the given window id (%d) is unknown", regs->ecx);
+                regs->eax = false;
+                break;    
+            }
+            } break;
```
29jm: Indentation is wrong + `return` in a void function. I also still don't like `being_dragged` being part of the wm_window state, but I can fix that after merging if needed. I can explain syscalls if you want, I know they are not the most straightforward concept.

modules/src/terminal.c
```diff
@@ -36,8 +37,18 @@ bool cursor = true;
...
+    .base_color = 0x00222221,
+    .highlight = 0x00030303,
+    .border_color = 0x00000000,
+    .text_color = 0x00FFFFFF,
+    .border_color2 = 0x00555555,
+};
+
 int main() {
     win = snow_open_window("Terminal", twidth, theight, WM_NORMAL);
+    win->color = &clr;
```
29jm: Again, the color scheme here is only for the titlebar, hardcoding it in the snow_draw_window function makes more sense than extending the state of window objects with a one time use variable.

ui/src/canvas.c
29jm: I mean, it should say "stop drawing"

modules/src/paint.c
```diff
@@ -57,11 +57,11 @@ int main(int argc, char* argv[]) {
         hbox_add(menu, W(cbutton));
     }
 
-    button_t* button = button_new("Clear");
+    button_t* button = button_new("Clear", NULL);
     button->on_click = on_clear_clicked;
     hbox_add(menu, W(button));
 
-    button_t* save_button = button_new("Save");
+    button_t* save_button = button_new("Save", NULL);
```
29jm: These second parameters should no longer be here.

29jm (commented):
+ PR does not currently build, has errors and warnings: run `make clean; make strict`, fix errors until it builds.
+ I feel like I'm insisting on stuff and I'm sorry for that. Please understand that clarity of code and concepts is very important to me, and I've learned that this can all go wrong very fast, and it takes a lot to correct.

Main issue here is that this is a big change. Main reason for that is probably the renaming of `rect_ŧ` wm-side to `wm_rect_t`. I think it would be far better to revert that change, making the PR easier to navigate for both of us.

mikumikudice:
I'll fix that tomorrow. Sorry, I think my last test before push had not been done with a ``make clean``.

kernel/src/sys/syscall.c
mikumikudice: I'm aware of what they do and for what they are used for. It's what is called when you use ``int <num>`` in assembly. But thanks, I think I forgot that I had to use the registers to return the value and ended thinking that I was in a normal function. My fault

mikumikudice:
My current local build, now I fixed every issue, works perfectly (as well as it's possible), including the save / clear buttons from paint. I'm just fixing some other minor stuff and then I'll push everything

kernel/include/kernel/wm.h
```diff
@@ -19,29 +19,33 @@
...
+    point_t pos;
     uint32_t id;
     uint32_t flags;
+    bool being_dragged;
     ringbuffer_t* events;
 } wm_window_t;
 
-// We exposed `wm_rect_t` to userspace, rename it here for convenience
-typedef wm_rect_t rect_t;
-
```
29jm: Please reinstate. It's very easy with VS Code:
+ re add typedef
+ go to rect.c, cursor on any wm_rect_t, F2 -> rename to rect_t
+ select all in wm.c/wm.h/rect.c to be renamed
+ fix typedef if it got renamed there too.

modules/src/terminal.c
```diff
@@ -138,19 +139,9 @@ int main() {
...
-    snow_draw_rect(win->fb, 0, 0, win->width, 20, 0x00222221);
-    snow_draw_border(win->fb, 0, 0, win->width, 20, 0x00000000);
-    snow_draw_string(win->fb, win->title, 4, 3, 0x00FFFFFF);
-    // border of the whole window
-    snow_draw_border(win->fb, 0, 0, win->width, win->height, 0x00555555);
-
-    /* Text content */
-
-    uint32_t y = 22; // below title bar
+    uint32_t y = UI_TB_HEIGHT + 4; // below title bar
```
29jm, line 144: This crashes the terminal when text reaches the bottom, e.g. running `ls` will crash it. Easiest fix is to set
```c
const uint32_t max_line = theight / char_height - 2; // from -1 to -2
```
The change to `margin` might have an effect, not sure.

29jm: Terminal is clunky, I know ^^ I was surprised how hard it is to manipulate text to make it fit a box, and I'm not even doing half the things I should be.

ui/include/ui.h
```diff
@@ -143,6 +157,7 @@ titlebar_t* titlebar_new(const char* title, const uint8_t* icon);
 void titlebar_set_title(titlebar_t* tb, const char* title);
 
 color_button_t* color_button_new(uint32_t color, uint32_t* to_set);
+void color_button_on_resize(color_button_t* button);
```
29jm: Prototype shouldn't be here. In color_button.c at most.

ui/src/button.c
```diff
@@ -45,6 +63,11 @@ void button_set_on_click(button_t* button, void (*callback)(button_t*)) {
     button->on_click = callback;
 }
 
+
+void button_set_on_release(button_t* button, void (*callback)(button_t*)) {
+    button->on_release = callback;
+}
+
```
29jm, line 70: Remove this function and associated `on_release` field. It's currently not exposed in the API (ui.h) and has no current use, or immediate uses I can think of for an UI button.

29jm: Conceptually I think I think clicks should be on mouse release (it's like this on pretty much every OS), mouse press and release should be different events handled as needed by widgets that need to make the distinction (e.g. canvas). This PR changes that in SnowflakeOS, clicks used to be on release. I will resolve that later, I should have thought of that much earlier.

29jm (changes requested):
Ok, please bear with me for those last changes, I promise they are the last ones I will ask for.

A minor issue I noticed also is that some gui apps no longer show their name in the titlebar, e.g. calc, files, while paint still does for some reason. Didn't debug, and will fix after merge unless you wish to investigate.

Another thing is that it hit me (see comment in button.c) that click events are now on mouse press and not mouse release (you mentioned that at some point I think and I didn't realize that implication) which is slightly weird, and probably slightly impractical. In the scope of this PR I don't mind too much, I will open an issue about this once merged as a reminder.

To your question about how to handle color schemes if the scheme isn't a field of `window_t`, I admit I hadn't thought that individual apps could ask to set their titlebar themes. It feels a bit like this should be global and user-set (don't you like some basic UI coherence across apps on your desktop?), and this can be done, but I don't know. If you don't mind let's drop that from the scope of this PR.

kernel/include/kernel/wm.h
29jm: > select all in wm.c/wm.h/rect.c to be renamed

And de-select other files or you will have a mess :)

ui/src/button.c
mikumikudice: I'll then make the call of the callback to the release, and let left the ``on_click`` to implement the button highlight. I'll try to implement it too.

mikumikudice:
I didn't entered GitHub since I pushed the last commit, sorry. I'll fix everything (hopefully the last time) tomorrow.

By the way I always say "tomorrow" because I always see the GitHub notifications on mobile at night

ui/src/button.c
29jm: I'm not sure I understand what you mean exactly here. `on_click` will stay put, right?

29jm:
> I didn't entered GitHub since I pushed the last commit, sorry. I'll fix everything (hopefully the last time) tomorrow.

Take the time you need :)

> By the way I always say "tomorrow" because I always see the GitHub notifications on mobile at night

We have like 7h of time zone lag I think!

mikumikudice:
I did everything, except fixing the title bar, but I know what's the problem (I thought it could happen but ended forgetting about that): the height-most UI child objects are covering the title bar text, that is being drawn ``UI_TB_HEIGHT`` pixels bellow the main position. Setting to all widgets to be drawn that amount of pixels bellow to will fix the problem (but I don't know how to fix it, so I'll put you in charge of this \:p). Again, thanks for the help and by accepting my changes

29jm:
Will merge this weekend (need to take some notes), and hopefully get to do some snowflake work while I'm at it as all of this gave me a few ideas. Thanks for your work and patience :)

29jm (approved):

modules/src/terminal.c
29jm: You did not fix this one :/

29jm:
Merged as promised :)

edit: made a few fixes on master, try them out!

